PR #298, opened 2018-11-30
abaplint bin command fix

sbcgua:
to #297

### DIFF
--- a/abaplint
+++ b/abaplint
@@ -1,3 +1,2 @@
-#!/usr/bin/env node
-
-require("./build/src/cli");
+#!/usr/bin/env sh
+node abaplint.js $@
--- a/abaplint
+++ b/abaplint
@@ -1,4 +1,3 @@
-#!/bin/sh
-":" //# comment; exec /usr/bin/env node "$0" "$@"
+#!/usr/bin/env node
 
 require("./build/src/cli");
--- a/abaplint
+++ b/abaplint
@@ -1,8 +1,2 @@
 #!/usr/bin/env node
-try {
-    require.resolve('worker_threads');
-    console.log('worker_threads found!');
-} catch(e) {
-    console.log('worker_threads NOT found');
-}
 require("./build/src/cli");
--- a/abaplint
+++ b/abaplint
@@ -1,2 +1,8 @@
-#!/usr/bin/env sh
-node abaplint.js $@
+#!/usr/bin/env node
+try {
+    require.resolve('worker_threads');
+    console.log('worker_threads found!');
+} catch(e) {
+    console.log('worker_threads NOT found');
+}
+require("./build/src/cli");
--- a/abaplint-experimental
+++ b/abaplint-experimental
@@ -1,3 +1,0 @@
-#!/usr/bin/env -S node --experimental-worker
-
-require("./build/src/cli");
--- a/abaplint-experimental
+++ b/abaplint-experimental
@@ -1,0 +1,3 @@
+#!/usr/bin/env -S node --experimental-worker
+
+require("./build/src/cli");
--- a/abaplint-x
+++ b/abaplint-x
@@ -1,0 +1,13 @@
+#!/usr/bin/env node
+const path = require('path');
+const { spawn } = require('child_process');
+
+const args = process.argv.slice(2);
+const abaplintPath = path.join(__dirname, 'abaplint');
+
+const pc = spawn('node', [
+  '--experimental-worker',
+  abaplintPath,
+  ...args ],
+  { stdio: 'inherit' }
+);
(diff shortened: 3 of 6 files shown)


larshp:
hmm, I changed it a while back, so that it is possible to add `--experimental-worker` in the executeable

sbcgua:
I see ...

Well, in my machine this works (abaplint file) - of course should be tested in the environment you want it to work ...

```
#!/usr/bin/env node --experimental-worker

const worker = require('worker_threads'); <<< TO CONFIRM FLAG IS ACTIVE

require("./build/src/cli");
```

Option 2: separate bin for experimental run e.g. `abaplint-beta` or `abaplint-with-node-params`

or .... ? just to ignore this issue

larshp:
doesnt seem to work, running on windows,

image

sbcgua:
Hmmm ... OK ... it is a bit controversial  at least at my level of shell knowledge.

the file above works if installed as package e.g.
```sh
npm link (or real npm install)
abaplint -h
```
If I understand correctly this is because npm actually wraps the script in a shell script which ignores shebang stuff in the file itself and ACCEPTS params
```sh
npm link
which abaplint
cat ... ^^^

#!/bin/sh                                                                              
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")                                  
                                                                                       
case `uname` in                                                                        
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;                                        
esac                                                                                   
                                                                                       
if [ -x "$basedir/node" ]; then                                                        
  "$basedir/node"  --experimental-worker "$basedir/node_modules/abaplint/abaplint" "$@"
  ret=$?                                                                               
else                                                                                   
  node  --experimental-worker "$basedir/node_modules/abaplint/abaplint" "$@"           
  ret=$?                                                                               
fi                                                                                     
exit $ret                                                                              

```

running `./xxxx --param` does not work indeed. However there is a param -S which allows this.

```
#!/usr/bin/env -S node --experimental-worker
```

Then I can run from dev dir directy
```
./abaplint -h
```

BUT this stops working with npm !!!

Imho to summarize: to add additional script with `-S`. If this is used for developer experiments then it should be just fine and for npm scripts use the one without additional params. (see last commit)

larshp:
also does not seem to work,
image

sbcgua:
Hmmmmm, will have a look

sbcgua:
OK, here is a new approach. Seems like `-S` is not available in all environments. Also I checked my older js projects how I did there and this is it. So the key fact is that shebang seems to be non universal.

1. abaplint renamed to abaplint.js (this is JS file isn't it ?)
2. bin in package.json refers to abaplint.js
3. js file keeps shebang. Seems like npm installs it differently in case of with and without shebang.
4. The above ensures `npm` related scenarion work (npm link, npm install), so AG scenario will also work (as it uses npm install)
5. Checked `npm link` + `abaplint -h` in irrelevant dir: works both for bash and cmd
6. added abaplint, 755, shell script, runs `node abaplint.js`. To run in dev env. `./abaplint -h` works
7. added abaplint.bat, cmd script, runs `node abaplint.js`. To run in dev env on win on cmd. 'abaplint -h' at dev dir works

larshp:
1: IMHO, no, it starts with `#!/usr/bin/env node` which is not JavaScript

sbcgua:
1: but it calls `require` :) Anyway the point was to make `abaplint` (the file w/o extension) a shell script and not a js file. We can keep `abaplint` name, just remove shebang from it and rename shell script to `abaplint.sh`

The final goal imho is:
- abaplint works after npm install (invoked by "user")
- abaplint can be called from dev dir for experiment with parameters (which is achieved by dedicated shell scripts for win/unix) (invoked by "developer")

And the problem is that it seems that it is not possible to do with one single file for all environments (at least from what I understand)

larshp:
any tips for reproducing the issue? see original issue

larshp:
what is the minimal solution to get it running, while ignoring the worker threads issue? Could just force users wanting to use the worker threads to do manual commands

sbcgua:
ok. attempt number 3 :) Have a look. (don't merge if OK, there a debug log output)

```
$ ./abaplint -h
worker_threads NOT found
Usage: abaplint [options] [file ...]
...
$ ./abaplint-x -h                   
worker_threads found!               
Usage: abaplint [options] [file ...]
...
```
after `npm link`
bash
image
cmd
image

larshp:
this looks good 👍 

the spawn and abaplint-x will work until/if worker threads are enabled by default

if you remove the debug output, then I'll merge it

sbcgua:
done
:tada: :smile:

larshp:
will release a new version tomorrow, if I remember it

larshp:
1.5.2 published, try again

sbcgua:
Works now :tada: 
image

PR #560, opened 2019-09-29
playground - travis rm -Rf old deploy before copying new

sbcgua:
remove old pg files before new deploy
P.S. `.git` folder stays untouched


larshp:
the CNAME file is required for the domain to work,
image

larshp:
easy solution is just to clean up the repo manually

sbcgua:
ah yes, missed that ... hmm ... how to make it better ... some kind of gitkeep ... will think about it but not today already :)

larshp:
just clean it up manually, its not something that changes much once the overall playground setup is done

PR #571, opened 2019-10-07
New Rule: No Dash in Form Names

### LINKED ISSUE #74: new rule: no dash in form name
FORM foo-bar.

g-back:
-fixes #74 
- adds .idea to gitignore (created with usage of phpstorm)

Question:
- Is your comment in the issue still relevant? Technically, your example ```DATA moo-boo TYPE c.``` is already shown as a parser_error.

### DIFF
--- a/src/rules/form_no_dash.ts
+++ b/src/rules/form_no_dash.ts
@@ -1,0 +1,55 @@
+import {ABAPRule} from "./_abap_rule";
+import {BasicRuleConfig} from "./_basic_rule_config";
+import {Issue} from "../issue";
+import {ABAPFile} from "../files";
+import * as Statements from "../abap/statements";
+import {Dash} from "../abap/tokens";
+
+
+/** Checks for a Dash in form names.*/
+export class FormNoDashConf extends BasicRuleConfig {
+}
+
+export class FormNoDash extends ABAPRule {
+
+  private conf = new FormNoDashConf();
+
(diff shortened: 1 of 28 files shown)


src/rules/form_no_dash.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  public runParsed(file: ABAPFile) {
+    const issues: Issue[] = [];
+
+    const struc = file.getStructure();
+    if (struc === undefined) {
+      return issues;
+    }
+    for (const form of struc.findAllStatements(Statements.Form)) {
+      if (form.concatTokens().includes(Dash.railroad())) {
```
larshp: I would like only to use the `railroad` method for visualization, suggest using `form.getTokens()` loop it and check `instanceof Dash`

larshp: alternatively, add a method `findAllTokens` similar to `findAllStatements` but for tokens

larshp:
Also please add test case for `FORM foo-.`, note that this is a DashW, "W" indicates that the dash is followed by whitespace

image

larshp:
all good

@g-back feel free to join the abapGit slack, https://abapgit-slackinviter.herokuapp.com/ there is a channel dedicated to abaplint

If you plan to make regular contributions, consider adding your name and affiliation in your profile

PR #659, opened 2019-12-04
Pretty-printer: lowercase non-keywords

### LINKED ISSUE #657: pretty printer should lower case non keywords
example, expect "STRING" to be lower cased,

image

FreHu:
Resolves #657


larshp:
nice, thanks

PR #681, opened 2019-12-12
New Rule: Unused Variables

g-back:
Draft: check unused variables in methods (part of #113)

- checks if a declared variable has been used as a `Source`
- todos:
  - inline data
  - field-symbols - special logic, usage as `Target` may be sufficient?

feel free to propose additional test cases or anything else I've missed


larshp:
its not this easy, but it can be a start

g-back:
> 
> its not this easy
could you elaborate on that?

larshp:
scoping rules, INCLUDEs, object attributes, method/procedure parameters

Long term I'd like to implement "find all references", and also checks for "variable is only written to", but not sure how yet, perhaps a specific walker for the AST, perhaps a directed graph like [LSIF](https://github.com/microsoft/language-server-protocol/blob/master/indexFormat/specification.md)

g-back:
Oh yeah absolutely, the topic as a whole is fairly complex, that's why I only check locally in the scope of methods. 
How would you like to proceed? Should we just wait until we have better methods to approach this topic or should we just start with this limited scope? 
If the latter: am I missing anything apart from field-symbols and inline data definitions?

larshp:
I think this code is okay, it will learn us more about the setup ABAP + abaplint + more, but be prepared that it might be removed / refactored

larshp:
spending some time on this today, lets see where it goes

larshp:
okay, now the syntax https://github.com/abaplint/abaplint/blob/master/src/abap/syntax/syntax.ts returns a `SpaghettiScope`, this contains all definitions + read + write information.

see https://en.wikipedia.org/wiki/Parent_pointer_tree

in a bit, the vscode extension will be able to highlight definitions + reads + write positions via https://github.com/abaplint/abaplint/blob/master/src/lsp/highlight.ts

larshp:
image

larshp:
closing

PR #876, opened 2020-05-08
ABAPObject type guard

marcellourbani:
Not that useful, but I find it cleaner.

This would have worked fine for the main class, but the symbol based one works for subclasses too
```typescript
  public static is(x:any):x is ABAPObject{
    return o instanceof ABAPObject;
  }
```
PS: this also worked fine, but leaving an unused error for [tag]. Lines 25 and 26 was added to make linters happy
```typescript
  public static is(x:any):x is ABAPObject{
    return !!x?.[tag];
  }
```

Could also have done this, but ignores any type error on that line, not only the unused variable one, but ignores any TS error for that line:

```typescript
  // @ts-ignore
  private [tag] = true;

  public static is(x:any):x is ABAPObject{
    return !!x?.[tag];
  }
```

### DIFF
--- a/packages/core/src/objects/_abap_object.ts
+++ b/packages/core/src/objects/_abap_object.ts
@@ -7,6 +7,7 @@ import {IClassImplementation} from "../abap/types/_class_implementation";
 import {IObject} from "./_iobject";
 import {Version} from "../version";
 
+const tag = Symbol("ABAPObject");
 export interface ITextElement {
   key: string;
   text: string;
@@ -14,10 +15,17 @@ export interface ITextElement {
 
 export abstract class ABAPObject extends AbstractObject {
   private parsed: readonly ABAPFile[];
+  private [tag] = true;
 
   public constructor(name: string) {
     super(name);
     this.parsed = [];
+  }
+
+  public static is(x:any):x is ABAPObject // silly overload to remove unused warning
+  // eslint-disable-next-line no-dupe-class-members
+  public static is(x:ABAPObject):x is ABAPObject{
+    return !!x?.[tag];
   }
 
   public parse(version: Version, globalMacros: readonly string[] | undefined): IObject {
--- a/packages/core/src/objects/_abap_object.ts
+++ b/packages/core/src/objects/_abap_object.ts
@@ -7,7 +7,6 @@ import {IClassImplementation} from "../abap/types/_class_implementation";
 import {IObject} from "./_iobject";
 import {Version} from "../version";
 
-const tag = Symbol("ABAPObject");
 export interface ITextElement {
   key: string;
   text: string;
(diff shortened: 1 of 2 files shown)


larshp:
what is the problem that will be solved? I'm lost

marcellourbani:
No problem really, just questionable typyngs:
Current get_abap.ts

```typescript 
export function getABAPObjects(reg: IRegistry): ABAPObject[] {
  return reg.getObjects().filter((obj) => { return obj instanceof ABAPObject; }) as ABAPObject[];
}
```

changed get_abap.ts:

```typescript 
export function getABAPObjects(reg: IRegistry): ABAPObject[] {
  return reg.getObjects().filter((obj) => { return obj instanceof ABAPObject; }) as ABAPObject[];
}
```

The former does't really check types, the compiler is perfectly happy with this:
```typescript 
export function getABAPObjects(reg: IRegistry): ABAPObject[] {
  // the cast will make the compiler happy no matter what the filter does
  return reg.getObjects().filter(()=>new Date().getTime() % 2 ) as ABAPObject[];
}
```
because the cast at the end is enforcing the type
A simpler change which has the same effect is this:

```typescript
// typeguard, validates the object type
const isABAPObject = (obj: any): obj is ABAPObject => obj instanceof ABAPObject;
// does have a similar problem,this would also make the compiler happy...
// but at least the type guard is defined on the object, not 
// const isABAPObject = (obj: any): obj is ABAPObject => new Date().getTime() % 2;

export function getABAPObjects(reg: IRegistry): ABAPObject[] {
  // no need for casting, the type guard guarantees the returned type
  return reg.getObjects().filter(isABAPObject);
}
```

larshp:
`const isABAPObject = (obj: any): obj is ABAPObject => obj instanceof ABAPObject;`
seems good

what are the differences vs the 
`return !!x?.[tag];`
solution? except for syntax

marcellourbani:
I think instanceof is fine.
I used the symbol because I remembered there was some quirk in instanceof but didn't remember which.

Looked it up and found [this](https://javascript.info/instanceof). Looks like the issues come when people mess with prototype chains.
A non problem unless annotations can influence it

Anyway both work fine:

```typescript
export class Foo {
  private [footag] = true

  public static is(x: any): x is Foo {
    return !!x?.[footag]
  }
}

export class Bar extends Foo {
  private [bartag] = true

  public static is(x: any): x is Bar {
    return !!x?.[bartag]
  }
}

const foo = new Foo()
const bar = new Bar()

console.log(`foo is of type Foo? ${foo instanceof Foo}`)
console.log(`foo instanceof Bar? ${foo instanceof Bar}`)
console.log(`foo is Foo? ${Foo.is(foo)}`)
console.log(`foo is Bar? ${Bar.is(foo)}`)
console.log(`bar instanceof Foo? ${bar instanceof Foo}`)
console.log(`bar instanceof Bar? ${bar instanceof Bar}`)
console.log(`bar is Foo? ${Foo.is(bar)}`)
console.log(`bar is Bar? ${Bar.is(bar)}`)
```
gives this output:
```
foo is of type Foo? true
foo instanceof Bar? false
foo is Foo? true
foo is Bar? false
bar instanceof Foo? true
bar instanceof Bar? true
bar is Foo? true
bar is Bar? true

```

marcellourbani:
Either way, I do think a well defined type guard is way better than a cast, even if works fine

marcellourbani:
PS: defining a function or an is method is pure style.
Symbol is probably more resilient but also more quirky

larshp:
yea, agree casts are bad, suggest doing the `const isABAPObject = (obj: any): obj is ABAPObject => obj instanceof ABAPObject;` approach

marcellourbani:
better?

larshp:
did you make a commit?

marcellourbani:
...and forgot to push... sorry

larshp:
push...

marcellourbani:
I did, [1b23665](https://github.com/abaplint/abaplint/pull/876/commits/1b236650c3b46730efd14d19b517663180ebe2b2)
Left it as ABAPObject.is rather than a function, partly out of lazyness and partly because I ike it better: one less import to deal with

PR #1112, opened 2020-07-31
add regression CI, wip

larshp:


larshp:
this will do for now, closes https://github.com/abaplint/abaplint/issues/1109

PR #1328, opened 2020-09-17
Add rule severity

FreHu:
Well this was annoying, but I don't see any other way than passing an extra parameter everywhere. 

Every rule now allows specifying severity `Error`, `Warning` or `Info`, defaulting to `Error`.

I have yet to modify the outputs.


larshp:
heh, yea, perhaps the `key` and `severity` can be combined into `this`, would probably need some other changes 🤷‍♂️

FreHu:
maybe something like `Issue.FromRule(rule, params)` and make params a union of all the variants.
`type params = RowRange | Position | Range | Token | ...`

larshp:
hmm

I'm okay with merging this as is

FreHu:
Yeah, I discovered that this is not as elegant as I thought. But I still need to adjust the outputs to have severity in them.

FreHu:
Hmm, now it's broken here, but it compiled for me after link local.

There also seems to be some kind of inconsistency with eslint, because marc's earlier commit passed and now locally I have errors about long lines:
image

larshp:
there is a problem with running the linter on CI, https://github.com/abaplint/abaplint/issues/1031

just add a eslint ignore

PR #1610, opened 2020-12-11
Fix default for local dependencies

mbtools:
`folder` and `files` are concatenated so it should be `/deps/**/*.*` as a default


larshp:
no, its under `/deps/src/**/*.*`,

image

larshp (changes requested):
see above

mbtools:
I see. I must have uploaded to /deps/*.* manually

mbtools:
just making a note where `loadDependencies` is located:
https://github.com/abaplint/abaplint/blob/master/packages/cli/src/cli.ts#L81

PR #1677, opened 2021-01-15
prefer_is_not add fix

jan-jezek:
prefer_is_not now supports quick fix!


jan-jezek:
test
before_fix
after_fix

Hi,

I have a situation where tests are working, but when I ran it on my code the fix is not applied correctly.
Can be reproduced in playground of this PR.

larshp:
try adding ENDIF. and/or a statement after the ENDIF.

```abap
IF NOT sdfdsf IS INITIAL.
ENDIF.
WRITE 'bar'.
```

jan-jezek:
You mean in test?
image
(I failed the test on purpuse ' is missing after bar)

Or in playground?
image

The same results as before.

larshp:
hmm, I'll take a look tomorrow

packages/core/src/rules/prefer_is_not.ts
```diff
@@ -2,8 +2,10 @@ import {Issue} from "../issue";
 import {ABAPRule} from "./_abap_rule";
 import * as Expressions from "../abap/2_statements/expressions";
 import {BasicRuleConfig} from "./_basic_rule_config";
+import {EditHelper, IEdit} from "../edit_helper";
 import {IRuleMetadata, RuleTag} from "./_irule";
 import {ABAPFile} from "../abap/abap_file";
+import {Position} from "..";
```
larshp, line 8:
```suggestion
import {Position} from "../position";
```

larshp:
also suggest moving the building of the fix to a new private method

jan-jezek:
The issue with tests running OK, but when applied on code it is not working properly still persists.

larshp:
yea, I'm guessing its related to the merging of the fixes

I might go ahead and merge the PR anyway

packages/core/src/rules/prefer_is_not.ts
```diff
@@ -50,13 +53,48 @@ IF NOT variable = 42.`,
...
-        issues.push(Issue.atToken(file, c.getFirstToken(), message, this.getMetadata().key, this.conf.severity));
+        const message = "Prefer expresion NOT to NOT expresion";
+
+        const fix = this.getFix(file, c);
+        if (fix === undefined) {
+          issues.push(Issue.atToken(file, c.getFirstToken(), message, this.getMetadata().key, this.conf.severity));
+        }
+        else {
+          issues.push(Issue.atToken(file, c.getFirstToken(), message, this.getMetadata().key, this.conf.severity, fix));
+        }
```
larshp, line 64: think this will work,
 
```suggestion
        issues.push(Issue.atToken(file, c.getFirstToken(), message, this.getMetadata().key, this.conf.severity, fix));
```

jan-jezek:
> yea, I'm guessing its related to the merging of the fixes
> 
> I might go ahead and merge the PR anyway

Is there a way how to it without merging the fixes?

packages/core/src/rules/prefer_is_not.ts
```diff
@@ -50,13 +53,43 @@ IF NOT variable = 42.`,
           continue;
         }
 
-        const message = "Prefer IS NOT to NOT IS";
-        issues.push(Issue.atToken(file, c.getFirstToken(), message, this.getMetadata().key, this.conf.severity));
+        const message = "Prefer expresion NOT to NOT expresion";
```
g-back, line 56: I think something is missing in the message (probably the "IS") - also it's "expression" (two 's')

jan-jezek: I don't like the message with IS, because the IS is not there most of the time.
The condition could look like IF NOT 5 = 2., IF NOT 5 IN range or IF NOT number BETWEEN 3 AND 5.
But I agree that expression (even without the grammatical mistake) is not clear as well.

larshp: I agree its not easy to understand, 

but keep it to the title in the style guilde, https://github.com/SAP/styleguides/blob/master/clean-abap/CleanABAP.md#prefer-is-not-to-not-is

jan-jezek:
> I might go ahead and merge the PR anyway

I don't really want this PR to be merged when I know it is not working in most (maybe all) cases.

larshp:
try changing the sequence, so the insert is done first, bottom to top, right to left is easier as it doesnt change the positions of the subsequent fixes

image

not

larshp (approved):
looks good to me, thanks 👍

larshp:
I'll release a new version sometime later this week

PR #1909, opened 2021-04-12
Codeframe formatter

### LINKED ISSUE #1908: New formatter - codeframe
inspired by https://eslint.org/docs/user-guide/formatters/#codeframe

Idea: more clear information from the output where is the issue.  Also will solve #986.

sbcgua:
fixes #1908

2021-04-12_18h18_44

A couple of thoughts

1) I added chalk module for coloring. Using it for other projects - lightweight and convenient. Disableable (see docs). It not strictly required for this feature but nice. And coloring can be used elsewhere in cli
2) Potentially (e.g. if extra modules are reeeeely unwanted) we can make this formatter a separate npm package (under abaplint umbrella). But I think this formatter is a good standard feature. I'd enable it for abapGit as well ...
3) The current formatter concept seems suboptimal to me. They create a long string and then writes it to the stream (memory inefficient). Why ? Why not to pass the stream (file or console) directly to formatter ? If OK, I can have a look on it.

some more stuff can be improved, though it is a separate issue. I'd use commander package for example for param parsing and also fix that progress output by default (it is a bit annoying, also because it is not visible in all console types)


packages/cli/package.json
```diff
@@ -57,5 +57,8 @@
         "webpack": "^5.28.0",
         "webpack-cli": "^4.6.0",
         "xml-js": "^1.6.11"
+    },
+    "dependencies": {
+        "chalk": "^4.1.0"
```
larshp, line 62: move it to `devDependencies`, everything is webpacked into one file in the cli, so no need for npm to install this as a dependency

larshp:
1: ok, no problem
2: -
3: are you thinking about performance? strings are easy to unit test

commander: hmm

progress: yea, that's a problem

packages/cli/package.json
larshp: ^just this comment, then its okay to merge 😄

sbcgua: hmmm ... that's another point which I would challenge a bit ... webpak is cool but very unintuitive. Why devDependency if it is a real dependency ? How to debug it if needed ? and etc ... honestly I'd suggest adjusting the devflow, bow it is not transparent. However this is a separate subject, maybe for slack :)

sbcgua:
> 3: are you thinking about performance? strings are easy to unit test

Yes. Well, so far no tests for formatters :) Ok, I agree, streams are a bit more complex to test, but not crucially. And in terms of architecture they are performant and universal. (hmmm, maybe we should check how eslint does it for a benchmark)

> commander: hmm

Why ? It is really awesome. Minimal descriptive configuration and you get a lot of param features and auto docs. Needs to get used to a bit because of event driven model, but it is not so big price. And dependency-less.

larshp:
commander: I think there was some story back long time ago, but I don't remember, suggest moving it to a separate issue

sbcgua:
yes, agree. The streams are also separate because all formatters have to be adapted.

So the only point remains is chalk in devDeps. Do we really need to move it ?

larshp:
yea

if adding it, it will just be an obsolete download each time `@abaplint/cli` is installed, can look into the webpack'ing, but suggest that for a separate issue

image

sbcgua:
Done. Agreed :)

P.S. wrote some side comments in slack

larshp (approved):

sbcgua:
P.S. streams: eslint also uses strings ... https://eslint.org/docs/developer-guide/working-with-custom-formatters
unexpected ... it is interesting why ...

PR #2228, opened 2021-11-06
update to node 16, set ignore-scripts

larshp:


larshp:
regression test failing is expected, the scripts differ

PR #2248, opened 2021-11-20
Enable hover for builtin methods

FreHu:
Adds the method hover also for builtin methods (not even adt knows this btw). I also reformatted all the builtin signatures.


larshp:
👍

larshp:
image

PR #2407, opened 2022-02-15
Bump up ABAP version to 7.56

jan-jezek:
ABAP 7.56 is already available to all customers:
https://help.sap.com/doc/abapdocu_latest_index_htm/latest/en-US/abennews-756.htm


larshp (approved):
cool 🚢

PR #2498, opened 2022-04-12
2.88.7, downport, VALUE with BASE

larshp:


larshp:
argh glob broken, need to roll back some stuff

PR #2615, opened 2022-08-01
JUnit Format - Stacktrace Format

lucasborin:
We are facing this issue:
https://github.com/jenkinsci/junit-plugin/issues/399

So, instead of using `space` for padding, I want to change it to `leading zeros`.


lucasborin:
Instead of:
image

it will use:
image

lucasborin:
@larshp: Would it be ok?

larshp:
no idea, I dont use it, but will merge

PR #2908, opened 2023-03-08
Fix "TABLE OF" for table types

mbtools:
Fix so first statement is syntax error:

```abap
DATA itab1 TYPE STANDARD TABLE string. "<<< should error
DATA itab2 TYPE STANDARD TABLE OF string.
```

Ref https://github.com/abaplint/abaplint/issues/2907


mbtools:
Now I see that the generic table types are mixed in with the full definition... Needs a bigger fix

larshp (approved):
looks good, thanks

PR #3139, opened 2023-10-23
implement_methods fixes

marcellourbani:
I think it's working ok, not too sure about the fix proposal.

Also, as this is about unknown interfaces, it will emit false positives when the target is not a method.


packages/core/test/rules/implement_methods.ts
```diff
@@ -523,4 +523,73 @@ ENDCLASS.`;
...
+        contents: `CLASS lcl DEFINITION.
+    PUBLIC SECTION.
+      ALIASES blah FOR zlif~blah.
+  ENDCLASS.
+  
+  CLASS lcl IMPLEMENTATION.
+  ENDCLASS.`,
+      },
+    ]);
+    expect(issues.length).to.equals(1);
```
larshp, line 539: hmm, should this be an error? `blah` might not be a method

marcellourbani: That's what I meant above. BTW, I thought I might have to check parent classes too to get this right

larshp:
🤷‍♂️ lets just merge it

larshp:
need to roll it back, due to regression,

[ABAP-Logger/src/zcl_logger.clas.abap:156](https://github.com/ABAP-Logger/ABAP-Logger/blob/main/src/zcl_logger.clas.abap#L156): Implement method "ty_symsg" from interface "zif_loggable_object"(implement_methods)
[ABAP-Logger/src/zcl_logger.clas.abap:156](https://github.com/ABAP-Logger/ABAP-Logger/blob/main/src/zcl_logger.clas.abap#L156): Implement method "ty_message" from interface "zif_loggable_object"(implement_methods)
[ABAP-Logger/src/zcl_logger.clas.abap:156](https://github.com/ABAP-Logger/ABAP-Logger/blob/main/src/zcl_logger.clas.abap#L156): Implement method "tty_messages" from interface "zif_loggable_object"(implement_methods)

marcellourbani:
Will look into it

PR #3340, opened 2024-06-15
Update getting_started.md

larshp:


larshp:
moving to https://github.com/abaplint/abaplint/pull/3341

PR #3721, opened 2025-10-01
Raise issue when abstract methods in non-abstract classes

fdupret:
Unrelated to https://github.com/abaplint/abaplint/pull/3717 but found when testing
We used to allow abstract methods in non-abstract classes. This PR enforces that abstract methods can only be declared in abstract classes. 
Error message is the one returned by SAP syntax check.


larshp (approved):
thanks

PR #3789, opened 2025-12-09
Added logic for renaming ICF services

cwildt:


packages/core/src/objects/rename/rename_icf_service.ts
```diff
@@ -0,0 +1,31 @@
...
+  public buildEdits(obj: IObject, oldName: string, newName: string): WorkspaceEdit | undefined {
+    if (!(obj instanceof ICFService)) {
+      throw new Error("RenameICFService, not a ICF Service");
+    }
+
+    let changes: (TextDocumentEdit | CreateFile | RenameFile | DeleteFile)[] = [];
+    const helper = new RenamerHelper(this.reg);
+    changes = changes.concat(helper.buildXMLFileEdits(obj, "ICF_NAME", oldName, newName));
+    changes = changes.concat(helper.buildXMLFileEdits(obj, "ORIG_NAME", oldName, newName));
+    changes = changes.concat(helper.buildXMLFileEdits(obj, "ICFHANDLER", oldName, newName));
```
larshp: not sure if the handler should be changed? at least not in the ICF renaming, but renaming a class should trigger ICF to be updated?

larshp:
thanks

PR #3799, opened 2025-12-14
fix: Preserve GUID suffix and update URL paths in SICF rename operations

cwildt:
## Summary
Enhances the SICF (Internet Communication Framework) service renaming logic to properly handle GUID suffixes in filenames and update URL path references.

## Changes
- **GUID Preservation**: SICF files with GUID suffixes (e.g., `z_svc 9def6c78d0beedf8d5b04ba6c.sicf.xml`) now preserve the GUID when renamed
- **URL Path Updates**: Added `buildURLFileEdits()` method to update URL paths within SICF XML files (e.g., `/sap/example/z_svc/` → `/sap/example/z_new/`)
- **Improved Object Lookup**: Enhanced object resolution to handle SICF services with GUID suffixes in their stored names
- **Backwards Compatibility**: Maintains support for legacy SICF files without GUIDs

## Modified Files
- `packages/core/src/objects/rename/rename_icf_service.ts` - Main renaming logic with GUID handling
- `packages/core/src/objects/rename/renamer_helper.ts` - New URL editing functionality
- `packages/core/src/objects/rename/renamer.ts` - Enhanced object lookup for SICF with GUIDs
- `packages/cli/test/rename.ts` - Updated integration tests
- `packages/core/test/objects/rename/rename_icf_service.ts` - Updated unit tests

## Testing
All tests updated to verify:
- Correct filename preservation with GUID suffix
- URL path replacement in XML content
- ICF_NAME and ORIG_NAME field updates
- Handler class reference updates

### DIFF
--- a/packages/cli/test/rename.ts
+++ b/packages/cli/test/rename.ts
@@ -123,11 +123,11 @@ ENDCLASS.`;
     expect(classNew).to.include("CLASS zcl_handler_new");
 
     // sicf file renamed and content updated (ICF_NAME and ICFHANDLER)
-    const sicfNew = mockFS.readFileSync("z_new 9def6c78d0beedf8d5b04ba6c.sicf.xml").toString();
-
-    expect(sicfNew).to.include("<ICF_NAME>Z_NEW</ICF_NAME>");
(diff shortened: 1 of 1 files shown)


larshp (approved):
thanks

PR #3844, opened 2026-02-20
fix(cds): add recursive parenthesized arithmetic with n-level nesting

abhisheksinha1601:
Introduce CDSArithParen as a mutual singleton with CDSArithmetics to enable unlimited nesting depth for parenthesized arithmetic expressions. Also fix decimal literals in CAST by ordering name before CDSInteger, and expose CDSArithParen in CDSElement for standalone grouped fields.


larshp (approved):
thanks

larshp:
published as 2.115.26

PR #3845, opened 2026-02-24
fix(cds): comprehensive parser fixes for SAP standard CDS views

abhisheksinha1601:
## Summary

This PR fixes a series of CDS parser bugs discovered by running SAP standard
views through the parser. Changes span the lexer and ~20 expression files.

### Lexer
- Flush token before entering string mode — fixes `else''` being tokenized as one token
- Treat `\t` (tab) as whitespace — fixes tab-indented SAP views silently failing
- Treat `@` as a token separator — fixes concatenated annotations like `@Foo: 'x'@Bar: 'y'`

### New syntax support
- `DEFINE ROOT ABSTRACT ENTITY` with association/composition fields (`cds_define_abstract`)
- `DEFINE VIEW Foo(A, B, C) AS ...` — column alias list (`cds_define_view`)
- `ASSOCIATION ... WITH DEFAULT FILTER` clause (`cds_association`)
- All join cardinality variants: `LEFT OUTER MANY TO EXACT ONE`, `INNER ONE TO MANY`, etc. (`cds_join`)
- Parenthesized join chains: `FROM (A CROSS JOIN B) LEFT OUTER JOIN C ON ...` (`cds_source`, `cds_join`)
- Hierarchy `DIRECTORY`, `ORPHANS`, `NODETYPE`, `CACHE FORCE/NONE/EMPTY` clauses (`cds_define_hierarchy`)
- Parameterized hierarchy `SOURCE` and `SIBLINGS ORDER BY` qualified names (`cds_define_hierarchy`)
- `$extension.*` wildcard field in element lists (`cds_element`)
- `FLTP_TO_DEC(input AS type)` function (`cds_function`)
- Decimal number literals e.g. `100.00` (`cds_integer`)
- Integer (including negative) parameter values e.g. `( P_Start: -11 )` (`cds_parameters_select`)
- Parameterized association path traversal in field lists: `src._Assoc(P_Key: val).Field` (`cds_prefixed_name`)
- Full path filter variants: `[cond]`, `[1:cond]`, `[1:inner]`, `[1:left outer where cond]` (`cds_prefixed_name`)

### Correctness fixes
- `cds_arithmetics`: unary-applied-to-paren form `-(TaxAmount)`
- `cds_annotation_simple`: support `null`, negative integers, `#('string')` enum values
- `cds_case`/`cds_cast`: correct operator precedence; arithmetic/paren subjects and values
- `cds_condition`: arithmetic on RHS of comparisons; `NOT (paren)` form
- `cds_function_input`, `cds_group_by`, `cds_cast`, `cds_element`: replace ad-hoc path logic with `CDSPrefixedName`
- `cds_string`: fix `abap.char 'X'` typed literal tokenization after lexer fix

### Tests
52 new test cases (629 lines), one per fixed syntax pattern. All existing 7700+ tests continue to pass.


larshp (approved):
thanks

PR #3850, opened 2026-03-03
fix(cds): comprehensive CDS grammar fixes for SAP standard views

abhisheksinha1601:
## Summary

This PR fixes a large number of CDS grammar issues discovered by running a parse-error loop against the full SAP standard CDS view corpus (~44k views). All fixes are covered by regression tests.

### Lexer (`cds_lexer.ts`, `cds_string.ts`)
- Tokenize `!` as a separate single-character token (fixes `Field!=Value` without spaces failing to parse)
- Fix block comment entry — consume `*` to prevent `/*/ ` from being treated as open+close
- Support backslash-escaped single quotes (`\'`) inside strings
- Normalize vertical tab (`\u000b`) to space in preprocessing

### Functions (`cds_function.ts`)
- Add `RATIO_OF(portion => X, total => Y)` function
- Add `CURR_TO_DECFLOAT_AMOUNT(amount)` function
- Add `TIMS_IS_VALID(time)` function
- Add `utcl_current()` zero-argument function (already in prior PR, confirmed working)

### Condition (`cds_condition.ts`)
- Add `CDSCast` as valid left-hand side of conditions (e.g. `cast(...) as type = value`)
- `ESCAPE` clause already restricted to `LIKE`/`NOT LIKE` only (resolved in rebase with #3849)

### Aggregate (`cds_aggregate.ts`)
- Support `avg(field AS type)` inline type coercion pattern
- Support `max('')` / `min('')` aggregate with string literal argument
- Use `CDSPrefixedName` to support path filters: `max(a._Assoc[filter].Field)`

### Annotations (`cds_annotation.ts`, `cds_annotation_simple.ts`)
- Support `#(dotted.path)` annotation values
- Support `@<` inline backward annotation syntax

### Arithmetics (`cds_arithmetics.ts`)
- Allow unary expressions as operands to handle `A + +B` and `A + -B`

### Association & Composition
- `cds_association.ts`: support text cardinality (`to exact one`, `to one`, `to many`)
- `cds_association.ts`: add `ASSOCIATION OF <cardinality> TO ...` form
- `cds_composition.ts`: support `composition of many` / `composition of one` bare forms
- `cds_composition.ts`: fix text cardinality (`one to many`, `exact one to many`)
- `cds_cardinality.ts`: fix text form to not include `OF` (handled by `CDSComposition`)

### View definitions
- `cds_define_view.ts`: make `DEFINE` keyword optional (deprecated `view ... as select` syntax)
- `cds_define_hierarchy.ts`: add `DEPTH <n>` clause; add `CACHE OFF`; support `MULTIPLE PARENTS NOT ALLOWED`
- `cds_define_projection.ts`: add `CDSParametersSelect` after source for `AS PROJECTION ON Src(P: val)`; add `TRANSIENT` keyword; support associations in projection body; support `CDSWithParameters` before `AS PROJECTION ON`
- `cds_define_table_function.ts`: allow annotations on return field definitions

### Aliases (`cds_as.ts`)
- Add greedy `opt(': type | LOCALIZED')` to support `as Alias : localized` pattern

### Join / Source
- `cds_join.ts`: add `INNER TO MANY/ONE/EXACT ONE` join type variants
- `cds_source.ts`: make parenthesized source self-recursive for arbitrary nesting depth
- `cds_select.ts`: trailing-comma tolerance in `{}` element lists; `WITH PARAMETERS` support

### Parameters
- `cds_with_parameters.ts`: allow trailing annotations after parameter type

## Tests

335 new lines in `test/cds/cds_parser.ts` covering all fixes above. Total: **7736 passing**.

### DIFF
--- a/packages/core/src/rules/definitions_top.ts
+++ b/packages/core/src/rules/definitions_top.ts
@@ -34,11 +34,11 @@ export class DefinitionsTop extends ABAPRule {
       shortDescription: `Checks that definitions are placed at the beginning of METHODs, FORMs and FUNCTIONs.`,
       extendedInformation: `https://docs.abapopenchecks.org/checks/17/`,
       tags: [RuleTag.SingleFile, RuleTag.Quickfix],
-      badExample: `FROM foo.
+      badExample: `FORM foo.
   WRITE 'hello'.
   DATA int TYPE i.
 ENDFORM.`,
-      goodExample: `FROM foo.
+      goodExample: `FORM foo.
   DATA int TYPE i.
   WRITE 'hello'.
 ENDFORM.`,


larshp:
@abhisheksinha1601 thanks, check above, some seem relevant, but I'm also okay with merging

abhisheksinha1601:
Addressed all 4 review comments in the latest commit (9f9113c):

1. **`cds_string.ts`** — Updated regex to `/^'(?:[^'\\]|''|\\'|\\(?\!'))*'$/` so standalone backslashes (e.g. `'C:\\temp'`) are accepted, while `\\'` is still treated as an escaped quote.

2. **`cds_association.ts`** — Added missing `TO` keyword in `ofForm`: `ASSOCIATION OF [0..1] TO Target ON ...`

3. **`cds_as.ts`** — Reordered `colonType` alternatives to `alt(CDSType, CDSName, "LOCALIZED")` so dotted/parameterised types like `abap.dec(21,6)` are matched fully before falling back to bare name.

4. **`cds_composition.ts`** — Replaced the pre-`OF` `opt(CDSCardinality)` with an inline `numericCardinality` (bracket form only), so text cardinality can only appear after `OF` as intended.

PR #3887, opened 2026-04-07
2.118.11: minor performance + bugfix

larshp:


larshp:
nice, looks like 10% better during statement categorization

PR #3942, opened 2026-05-18
bugfix: allow CASE expression inside CAST in SELECT field list

pg-skoll:
Previously, the SQL CAST expression only accepted SQLFunctionInput as its inner expression, which prevented valid ABAP SQL like:

  SELECT CAST( CASE field WHEN 'X' THEN ' ' ELSE 'X' END AS CHAR ) AS col
    FROM ztab INTO TABLE @DATA(lt).

From being parsed correctly.

This change updates the `cast` constant in sql_function.ts to use `altPrio(SQLCase, SQLFunctionInput)`, allowing either a CASE expression or a regular function input inside CAST. Two new test cases are added in the SELECT statement tests to cover both simple and correlated CASE expressions inside CAST.


larshp (approved):
thanks, will be part of next release

dunno when I'll release the next version, perhaps today or in a few days

larshp:
published as 2.119.21

PR #3975, opened 2026-06-10
ABAP SQL parser: support for PRIVILEGED ACCESS and OPTIONS clause

### LINKED ISSUE #3889: SELECT + WITH PRIVILEGED ACCESS

cohomology:
The OPTIONS clause is supported on SELECT (single and multi-row), SELECT loop, and OPEN CURSOR statements.

Closes #3889.


cohomology:
I’m struggling to understand the intended meaning of Version.Cloud in abaplint's versioning scheme.

Is it meant to represent:

a.) A language version (similar to ABAP language versions such as v0, v2, v5), i.e. a property of the program?

or

b.) The ABAP version available in the Steampunk/ABAP Cloud runtime, i.e. a property of the runtime or kernel?

To me, both interpretations raise some concerns.

For (a), it seems problematic because the ABAP language version and the ABAP/kernel release are orthogonal concepts. Combining them in a single enumeration makes the model harder to reason about and leads to inconsistencies. Also, it is not clear whether ABAP v2 (key users) or ABAP v5 is meant. Language constructs which are NEW in ABAP Cloud but not released for customers yet (i.e. are only used by SAP in customer visible code) could not be added. The main problem, if course, is, that Version.Cloud would neither be newer, nor older than, say 758.

For (b), Version.Cloud appears to become a moving target. For example, if a syntax construct is guarded by Version.Cloud, that guard may eventually need to be replaced with a specific on-premise release once the feature becomes available there as well. In practice, such updates are unlikely to occur consistently, so version guards may gradually lose their original meaning over time.

Would it be more robust to model this as a tuple (release, language_version)?

In that model:

release would represent the platform release, including cloud releases such as 917, 918, etc.
language_version would represent the ABAP language version (v0, v2, v5, ...), and possibly legacy variants such as non-Unicode if those need to be supported.

The comparison logic could then handle the fact that cloud and on-premise releases belong to different release tracks (for example, 816 being newer than 917 in some contexts).

Am I misunderstanding the purpose of Version.Cloud, or is there a rationale behind the current design that I am missing?

larshp:
Version.Cloud: yea, its a mess, I added it before knowing how SAP would do the versioning

https://github.com/abaplint/abaplint/issues/3976 opened

larshp (approved):
thanks

PR #3980, opened 2026-06-11
prefix_is_current_class: skip members shadowed by parameters or locals

### LINKED ISSUE #3707: prefix_is_current_class bug: Narrow case of necessary reference to the class
Narrow case of necessary reference to the class

- https://github.com/sbcgua/agapi/blob/master/src/zcx_agapi_error.clas.abap#L88
- https://github.com/sbcgua/agapi/actions/runs/17973562612/job/51121732668

```abap
  begin of ls_split,
    v1 like zcx_agapi_error=>v1,
```

Where:

- the method is static
- and there is a `v1` method param (so reverence to pure v1 is ambiguous

### LINKED ISSUE #3755: prefix_is_current_class: false positive
Case where class-name is required but abaplint shows error. Often found in constructors.

```
/zfoo.prog.abap[13, 5]: Reference to current class can be omitted: "ZCL_TESTS=>"(prefix_is_current_class)
```

Example:

```abap
CLASS zcl_tests DEFINITION PUBLIC.
  PUBLIC SECTION.
    "! Parameter with same name as class-data
    CLASS-METHODS init
      IMPORTING
        test TYPE i.
  PRIVATE SECTION.
    CLASS-DATA test TYPE i.
ENDCLASS.

CLASS zcl_tests IMPLEMENTATION.
  METHOD init.
    zcl_tests=>test = test.  "<<< can't remove class here
  ENDMETHOD.
ENDCLASS.
```

Similar to #3707 and #1952.

palimkarakshay:
Fixes #3755, fixes #3707

When the member referenced via `current_class=>member` is shadowed by a method parameter or a local declaration with the same name, the prefix is required — removing it (or applying the quickfix) changes the meaning of the code:

```abap
CLASS zcl_tests IMPLEMENTATION.
  METHOD init.
    zcl_tests=>test = test.  " prefix cannot be removed, "test" is also an importing parameter
  ENDMETHOD.
ENDCLASS.
```

### What this does

In `checkClasses`, for statements inside method implementations, the rule now builds a set of names that shadow class members in that method:

- parameter names from the matching `METHODS`/`CLASS-METHODS` definition in the same file (importing/exporting/changing/returning),
- local declarations in the method body (`DATA`/`CONSTANTS`/`STATICS`/`TYPES` via `DefinitionName`, plus inline `DATA(...)`/`FINAL(...)`).

A `className=>member` reference whose member is in that set is skipped. If a statement contains several references, the first non-shadowed one is still reported (previously: always the first occurrence).

The rule stays single-file/syntactic, so parameters of interface methods or redefinitions declared in other files are not seen — behaviour for those stays as before.

Tests cover both issue examples (param shadowing assignment + `LIKE` reference in `BEGIN OF`), local and inline `DATA` shadowing, and two guards (different-name param still reported, mixed statement still reported).


larshp:
@palimkarakshay thanks, can you help signing the CLA ?

larshp (approved):
thanks

PR #3981, opened 2026-06-12
ABAP SQL parser: complete UNION support (parenthized groups)

cohomology:
This adds support for A UNION ( B UNION C ) and other combinations.


packages/core/test/abap/expressions/sql_set_op.ts
```diff
@@ -0,0 +1,505 @@
...
+}
+
+describe("Test expression, SQLSetOp", () => {
+
+  it("UNION SELECT: INTO is at top-level Select", () => {
+    const stmt = parseSelect("SELECT zcol FROM ztab UNION SELECT zcol FROM ztab2 INTO TABLE @DATA(res)");
+    assertASTEqual(
+      ["Select", {}, [
+        ["Select", {}, [
+          ["Identifier", {text: "SELECT"}, []],
```
larshp, line 21: hmm, I dont like to have many tests like this, they make it harder to refactor the implementation

packages/core/src/abap/2_statements/expressions/select_core.ts
```diff
@@ -0,0 +1,34 @@
...
+import {SQLFieldList, SQLFrom, SQLCond, SQLSource, SQLClient, DatabaseConnection,
+        SQLOrderBy, SQLHaving, SQLForAllEntries, SQLHints, SQLFields,
+        SQLIntoList, SQLOptions, SQLPrivilegedAccess} from ".";
+import {Version} from "../../../version";
+import {IStatementRunnable} from "../statement_runnable";
+import {SQLGroupBy} from "./sql_group_by";
+import {SQLIntoStructure} from "./sql_into_structure";
+import {SQLUpTo} from "./sql_up_to";
+
+export function buildSelectCore(into?: IStatementRunnable, allowOrderBy = true): IStatementRunnable {
```
larshp, line 11: this file is special, suggest renaming, prefix it with underscore "_select_core.ts" or make it a real expression

packages/core/test/abap/expressions/sql_set_op.ts
cohomology: What to do about that then? If we want people to actually *use* the AST, we should test it. 
For refactoring, we have AI nowadays ;-). Just a couple of tokens ...

packages/core/src/abap/2_statements/expressions/select_core.ts
cohomology: I will do that in my next commit

packages/core/test/abap/expressions/sql_set_op.ts
larshp: yea..

I dont consider the AST a released feature as such, if it changes and somebody uses it its their problem. The AST can change at any time as part of a patch release

cohomology: When abaplint's parser is feature complete, there is no need to change the AST frequently and you can release it as official feature after testing.

When I was working on a similar parser a while back, a colleague introduced these kinds of AST tests. I was strictly opposed to it for about a year and only gave up after a few thousand tests existed. I am now convinced that they are really great if you have built some tooling around it.

cohomology: There where actually some serious problems during development of this commit, for example the INTO clause was consumed by some subselect. Found it with this test.

PR #3983, opened 2026-06-14
ABAP SQL parser: major refactoring

cohomology:
- Dramatically improve performance by removing all per() constructs in SELECT. SELECT has a "well defined" and "easy" exact ordering which can be used (some irony here). Now also very large ABAP SQL statements can be parsed quickly (see with_large.ts, = 52k WITH statement).

- Distinguishing between Select and Select Loop is nearly impossible for the parser, as it can depend upon different things at different places being together in the same statement. It is also slow doing so. Better approach: Parsing the SELECT and distinguishing between Loop and non-loop by an afterburner.

- All kinds of bugfixes to expressions and common table expressions, for example CTE interface lists.


packages/core/src/abap/2_statements/statement_parser.ts
```diff
@@ -289,13 +384,38 @@ export class StatementParser {
...
-        return new StatementNode(st.statement, statement.getColon(), pragmas).setChildren(match);
+        const matched = new StatementNode(st.statement, statement.getColon(), pragmas).setChildren(match);
+        return this.reclassifySelect(matched, st.statement, pragmas);
       }
     }
 
     return statement;
   }
 
+  private reclassifySelect(
```
larshp: extract this part to new file? along with the utility methods

cohomology: I will do that.

cohomology:
Ok like this?

packages/core/src/abap/2_statements/statement_parser.ts
```diff
@@ -14,8 +14,9 @@ import {ExpandMacros} from "./expand_macros";
 import {Pragma} from "../1_lexer/tokens";
 import {IRegistry} from "../../_iregistry";
 import {IStatementRunnable} from "./statement_runnable";
+import {reclassifySelect} from "./expressions/_select_reclassify";
```
larshp: yea, perhaps move the _select_reclassify.ts file to the same folder as statement_parser.ts, then we know its not a n expression

cohomology: Done

PR #3995, opened 2026-06-15
ABAP SQL parser: minor improvements

cohomology:
- BYPASSING BUFFER in all positional variants
- CLIENT SPECIFIED with column list / dynamic
- GROUP BY arbitrary expressions
- Unary minus in field list and CASE THEN/ELSE
- OFFSET after INTO UP TO
- %_HINTS accepts host variables and field chains
- Multiple db hints for the same db vendor
- Dynamic field offset/length notation inside Dynamic
- Parenthesized THEN/ELSE values in CASE
- CONNECTION in all INTO-before-FROM and SINGLE positions
- INTO wa UP TO N ROWS FOR ALL ENTRIES ordering
- Arbitrary nesting of JOINs

### DIFF
--- a/packages/core/src/abap/2_statements/expressions/_select_core.ts
+++ b/packages/core/src/abap/2_statements/expressions/_select_core.ts
@@ -45,30 +45,32 @@ export function buildSelectCore(into?: IStatementRunnable, allowOrderBy = true):
     ));
   }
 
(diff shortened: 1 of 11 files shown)


larshp:
@cohomology thanks, you should have access to create branches in this repository, this will trigger the regression test(it is not triggered from forks)

PR #4003, opened 2026-06-16
Add dynamic structure and reference chains ref->(bla)-(blub)->*

cohomology:


packages/core/src/abap/2_statements/expressions/_dynamic_access.ts
```diff
@@ -0,0 +1,18 @@
...
+const dynClose = altPrio(tok(ParenRight), tok(ParenRightW));
+const dynInside = altPrio(new FieldChain(), Constant);
+
+export function dynComp(...inside: IStatementRunnable[]): IStatementRunnable {
+  return ver(Version.v757, seq(tok(Dash), dynIdent, dynInside, ...inside, dynClose));
+}
+
+export function dynAttr(): IStatementRunnable {
+  return ver(Version.v757, seq(tok(InstanceArrow), dynIdent, dynInside, dynClose));
+}
```
larshp, line 18: might want/need these as expressions in the future, but its good for now

PR #4029, opened 2026-06-23
selection_screen_texts_missing: handle NO-DISPLAY and inline cases

fdupret:
There are some cases where it's OK not to have a text assigned to a parameter:
- with the NO-DISPLAY addition
- When the parameter is inside a `BEGIN/END OF LINE` statement (the selection text is not rendered on screen)


larshp (approved):
thanks

PR #4037, opened 2026-06-24
Fix abaplint usage on Windows on a disk other than C:

### LINKED ISSUE #4036: abaplint cli won't work on Windows when running on another disk than C:
When running abaplint on a repo located on D:\ :
```
generic[1, 1] - Error: spawnSync C:\WINDOWS\system32\cmd.exe ENOENT at Object.spawnSync (node:internal/child_process:1127:20) (generic_error) [E]
```
Moving the repo from D: to C: fixes the issue.

fdupret:
Fixes #4036
Had three cascading issues:
- git clone would fail
- abaplint later failed to find the cloned dependencies (`Error: No files found, /Users/xxx/AppData/Local/Temp/abaplint-EYaTVJ/src/**/*.* undefined (generic_error) [E]`)
- `abaplint --rename` assumed we were always working on C:

Tested locally.
Disclaimer: I used Claude to fix this and don't master 100% of the code, feedback welcome

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -113,14 +113,13 @@ async function loadDependencies(config: Config, compress: boolean | undefined, b
 
     if (d.url) {
       process.stderr.write("Clone: " + d.url + "\n");
-      let dir = fs.mkdtempSync(path.join(os.tmpdir(), "abaplint-"));
-      dir = FileOperations.toUnixPath(dir);
+      const dir = fs.mkdtempSync(path.join(os.tmpdir(), "abaplint-"));
       let branch = "";
       if (d.branch) {
         branch = "-b " + d.branch + " ";
       }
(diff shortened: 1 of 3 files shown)


packages/cli/src/index.ts
```diff
@@ -113,14 +113,16 @@ async function loadDependencies(config: Config, compress: boolean | undefined, b
 
     if (d.url) {
       process.stderr.write("Clone: " + d.url + "\n");
-      let dir = fs.mkdtempSync(path.join(os.tmpdir(), "abaplint-"));
-      dir = FileOperations.toUnixPath(dir);
```
larshp, line 117: can the logic be added inside the toUnixPath method? its used from multiple places if I remember correctly

fdupret: Done. It was used by `--rename`, which also didn't work outside C:. Now confirmed to be working after 4c7fc589af26b7b6145b71a2835a59e6480d3cd4

larshp:
the filenames is a mess, will merge, and then see what happens

PR #4061, opened 2026-06-30
Prevent infinite recursion in macros

cohomology:


larshp:
heh, always another ABAP feature, always guessed this was not possible

anyhow abaplint and such does macro expansion on compile time, guess its okay, nobody should use macros in 2026 IMHO

cohomology:
> heh, always another ABAP feature, always guessed this was not possible
> 
> anyhow abaplint and such does macro expansion on compile time, guess its okay, nobody should use macros in 2026 IMHO

Agents love to create Macros in 2026 :-(.

larshp:
> > heh, always another ABAP feature, always guessed this was not possible
> > anyhow abaplint and such does macro expansion on compile time, guess its okay, nobody should use macros in 2026 IMHO
> 
> Agents love to create Macros in 2026 :-(.

ugh, I've not seen that, was looking for a existing rule to not allow macros, but cannot find it, adding

PR #4085, opened 2026-07-05
ABAP/CDS parser: fix 16 CDS parser gaps; fix ABAP parser crash

cohomology:
Multiple VALUE #( FOR x IN tab WHERE ( field = ref->* ) ( x ) ) parameters in a single method call caused exponential backtracking.

All kinds of CDS parser improvements from real world CDS examples.


larshp:
@cohomology something broke, see the regression test results above

cohomology:
> @cohomology something broke, see the regression test results above

Ok, look into it.

PR #4096, opened 2026-07-08
CDS Parser: Support Quoted Identifiers + Backwards Annotations in CDS

cohomology:
They are removed when saving, but CDS officially supports and parses them.


larshp (approved):
thanks

PR #4128, opened 2026-07-16
Fix main file lookup for URL-encoded namespaces

normanre:
## Summary

- Recognize URL-encoded namespace separators (`%23`) when resolving an ABAP object's main file.
- Prevent a class test include from being selected as the main class when file enumeration returns it first.
- Add a regression test where a namespaced test-class include precedes the main class file.

## Problem

VS Code represents namespaced filenames such as:

`#namesp#cl_foobar.clas.abap`

as URI-encoded paths:

`%23namesp%23cl_foobar.clas.abap`

`getMainABAPFile()` only searched for the `#namesp#` form. When that search failed, it returned the first ABAP file as a fallback. Since filesystem enumeration order is not deterministic, this could select `.clas.testclasses.abap` and resolve the global class as its local test class.

Related issue: https://github.com/abaplint/vscode-abaplint/issues/201

### DIFF
--- a/packages/core/src/objects/_abap_object.ts
+++ b/packages/core/src/objects/_abap_object.ts
@@ -80,10 +80,10 @@ export abstract class ABAPObject extends AbstractObject {
   }
 
   public getMainABAPFile(): ABAPFile | undefined {
-    // todo, uris, https://github.com/abaplint/abaplint/issues/673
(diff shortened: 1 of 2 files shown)


larshp:
thanks

PR #4135, opened 2026-07-22
Parse FUNCTION statements with inline parameter signatures

abhisheksinha1601:
## Summary
- The `FUNCTION` statement grammar only matched `FUNCTION <name>.`, so function modules with an inline parameter signature (`FUNCTION name IMPORTING VALUE(...) TYPE ... EXPORTING ...`) parsed as `Unknown` and the `FUNCTION...ENDFUNCTION` structure never built.
- Adds a `FunctionModuleParameters` expression family for the definition-side signature, mirroring the existing call-side `FunctionParameters` container (single container of `optPrio` clauses in canonical order: IMPORTING, EXPORTING, CHANGING, TABLES, RAISING/EXCEPTIONS).
- Per-parameter grammar is dedicated (not reused from `MethodParam` or `FormParam`) because classic FM params can be type-less and can include a `TABLES ... STRUCTURE ...` clause that methods don't have.
- `RAISING`/`EXCEPTIONS` are reused directly from the existing method-def expressions.
- The `verNotLang(LanguageVersion.KeyUser, ...)` restriction on the statement is unchanged.

## Test plan
- [x] `npx tsc` — clean compile
- [x] `npm test` (mocha + eslint + schema + api-extractor) — 10561 passing, 0 failing
- [x] New `test/abap/statements/function_module.ts` — 27 cases covering IMPORTING/EXPORTING/CHANGING/TABLES (incl. `STRUCTURE`), type-less params, `OPTIONAL`, `DEFAULT`, `LIKE`, `EXCEPTIONS`, `RAISING`/`RESUMABLE`, and combined signatures
- [x] New `test/abap/structures/function_module.ts` — confirms `FUNCTION...ENDFUNCTION` now builds for inline-signature FMs


larshp:
suggest fixing abapGit first, it doesn't serialize function modules this way

abhisheksinha1601:
Thanks for the context — you're right that abapGit never serializes function modules this way; the signature always goes through the `*"Local Interface:` comment block + the `.fugr.xml`, and abaplint's own test fixture (`fugr_import_export.ts`) confirms that's the canonical form.

The FM source triggering this isn't from an abapGit-serialized repo, though — it comes from a direct SAP ADT source fetch (`GET .../functions/groups/{FG}/fmodules/{FM}/source/main`), which returns the FM body with the parameter signature inline on the `FUNCTION` statement rather than as a comment block. That's a different (but common) ingestion path: tooling that talks to ADT directly instead of going through abapGit's serialization will hit this same parse failure on any FM with this shape.

So this PR isn't proposing that abapGit's format is wrong or should change — it's making abaplint's `FUNCTION` grammar tolerant of a source shape that real SAP systems do produce outside of abapGit's own serialization. It's purely additive: the standard `FUNCTION name.` + comment-block form is untouched and still parses exactly as before (covered by the existing tests), this just adds acceptance of the inline form as an alternative rather than failing it as `Unknown`.

Happy to narrow the scope or gate this differently if that addresses the concern — let me know what would make this easier to land.

abhisheksinha1601:
Dug into this a bit — checked zcl_abapgit_object_fugr (and the inherited strip_generation_comments in zcl_abapgit_objects_program) to see what abapGit actually does with FM source.

Turns out there's nothing to fix on abapGit's side: serialize_functions calls SAP's own RPY_FUNCTIONMODULE_READ_NEW, takes whatever source/new_source comes back, strips two unrelated auto-generated header comments (#**regenerated at *, #**generation date:*/#**generator version:*), and writes the rest to the .abap file verbatim. It never parses or reformats the interface region — the comment-block form we see in abapGit's own repo (e.g. zabapgit_parallel.fugr.z_abapgit_serialize_parallel.abap) is just what SAP's Function Builder happens to return for a normally-generated FM, not something abapGit imposes.

So abapGit is already "tolerant" of any interface shape by construction — it doesn't look at it. If a system's RPY_FUNCTIONMODULE_READ_NEW ever does return an inline-signature FM (older/legacy function pools, hand-edited sources, etc.), abapGit would serialize that verbatim too, and this same abaplint gap would surface downstream.

Worth noting abapGit's own package.json depends on @abaplint/cli/@abaplint/runtime/@abaplint/transpiler-cli for its CI lint and transpilation — so this fix helps abapGit's own pipeline too, for free, whenever it bumps its abaplint dependency.

larshp:
I'd like to move forward, improving abapGit and abaplint, one step at a time. Moving abapGit to work on any stack, plus using the ADT endpoints too, but time is low, I also need to try paying my rent.

Pull requests in abapGit is welcome, this PR is good, but first things first, or we can wait until somebody else implements it, or I have time.

Lets work towards improving the ecosystem, not segregating it.

larshp:
closing, welcome after abapGit works with/for the code

PR #4149, opened 2026-07-28
fix: urls to external sites

mbtools:
I updated broken and outdated URLs. I'm not sure if you want to keep the check script in the repo.

Not changed:

- Slack invite: Doesn't work anymore (also abapGit readme). The inviter.co replacement is free and works well but you have to register and config a landing page again
- rule_page.ts contains https://schema.abaplint.org/dummy.json (404). Not sure if this should be https://schema.abaplint.org/schema.json


larshp (approved):
thanks

PR #4237, opened 2026-08-23
Add XML BOM rule

mbtools:
Adds xml_bom with a quick fix and tests.

PS: a bit of an experiment. I burned some tokens using codex on my iPhone with gpt terra medium.


larshp (approved):
thanks

PR #4242, opened 2026-08-23
line_break_style: also check XML files

oblomov-dev:
The rule's own shortDescription names abapGit as the reason: "abapGit does not work with CRLF". But run() only ever looked at files ending in .abap, so a .clas.xml, .intf.xml or any other abapGit serialization sidecar with CRLF passed unnoticed. Measured on 2.120.33 with line_break_style explicitly enabled: a CRLF .intf.xml produced zero findings. That is the same defect the rule was written for, just on the other half of what abapGit serializes.

The filter now accepts .xml alongside .abap, and MIME objects are skipped the same way whitespace_end skips them, with the same extendedInformation: a .w3mi.data.xml is content rather than a serialization and may legitimately contain CRLF.

Title, shortDescription and the generated scripts/schema.json are updated to match the widened scope.

### DIFF
--- a/packages/core/scripts/schema.json
+++ b/packages/core/scripts/schema.json
@@ -2761,7 +2761,7 @@
                   "type": "boolean"
                 }
               ],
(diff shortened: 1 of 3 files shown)


larshp:
thanks

PR #4243, opened 2026-08-24
feat: add support for compositions in cds_field_order rule tests

fdupret:
Compositions are now treated like associations in `cds_field_order`.

Previously only `association` aliases were collected, so a `composition [0..*] of ... as _Comp`
listed at the end of the field list was seen as a regular field — producing a false "Associations must be listed last" finding on every field after it.

`getAssociationNames` now also scans `CDSComposition` nodes.


larshp (approved):
thanks

PR #4291, opened 2026-09-14
implement_methods: honour DEFAULT IGNORE and DEFAULT FAIL

oisee:
An interface method declared `DEFAULT IGNORE` or `DEFAULT FAIL` need not be implemented by an implementing class, but the rule asks for one anyway:

```abap
INTERFACE zif_t PUBLIC.
  METHODS required.
  METHODS optional DEFAULT IGNORE.
ENDINTERFACE.
```

with a class implementing only `required` gives

```
Implement method "optional" from interface "zif_t" [E]
```

The control is that **removing `DEFAULT IGNORE` gives exactly the same message**, so the rule was not weighing the modifier, it was not seeing it. The parser has understood the syntax since 7.40 SP08 (`method_def.ts`), but `InfoMethodDefinition` did not carry it, so there was nothing for the rule to read.

Two fields added beside `isForTesting` and `isFinal`, filled the same way from the statement's own tokens, and the rule skips a method that has either. `DEFAULT FAIL` is included because it is equally optional to implement — the difference is only what happens when an unimplemented one is called, which this rule has no view on.

Five tests. The three that fail without the change are the two modifiers and the neighbour case: a method beside a `DEFAULT IGNORE` one is still required. The other two are guards — without the modifier the message must still appear, and implementing such a method anyway must still be fine.

Suite 10977 passing before and 10982 after, none failing either way, lint clean.

Found on a repository that legitimately omits an optional method and could therefore neither be linted nor transpiled.

A note on where this comes from: this is a fork PR because I have no push rights here, so `regression.yml` skips it by its own `github.repository == 'abaplint/abaplint'` condition. Happy to run the regression locally and paste the result if that would help.

https://claude.ai/code/session_01GU3xTpNL1QDbNkGzuZ4pqg


larshp (approved):
thanks

PR #4298, opened 2026-09-16
Downport: strip WITHOUT AUTHORITY-CHECK from CALL TRANSACTION

### LINKED ISSUE #4297: `downport --fix` (syntax.version v702) does not strip `WITHOUT AUTHORITY-CHECK` from `CALL TRANSACTION`, even though the clause isn't valid on older targets
### Summary
When running abaplint's downport auto-fix (`rules.downport: true`, `syntax.version: "v702"`) against source containing `CALL TRANSACTION ... WITHOUT AUTHORITY-CHECK ...`, the `WITHOUT AUTHORITY-CHECK` addition is left untouched. On our v702 target system this statement fails at runtime/activation, so every downport currently needs a manual post-processing step (a `sed` pass in CI) to strip it.

### Example
Input (valid 7.5x, our source style):
```abap
CALL TRANSACTION 'PA20' WITHOUT AUTHORITY-CHECK AND SKIP FIRST SCREEN.
```

Expected after `abaplint --fix` with a v702 target:
```abap
CALL TRANSACTION 'PA20' AND SKIP FIRST SCREEN.
```

Actual: statement is left unchanged.

### Minimal reproducible example

`abaplint.json`:
```json
{
  "global": {
    "files": "/src/**/*.*"
  },
  "syntax": {
    "version": "v702"
  },
  "rules": {
    "downport": true
  }
}
```

`src/zfoo.prog.abap` (before):
```abap
REPORT zfoo.

DATA(lv_foo) = 1.

CALL TRANSACTION 'PA20' WITHOUT AUTHORITY-CHECK AND SKIP FIRST SCREEN.
```

Run:
```
npx abaplint abaplint.json --fix
```

`src/zfoo.prog.abap` (after — actual output, abaplint 2.119.48):
```abap
REPORT zfoo.

DATA lv_foo TYPE i.
lv_foo = 1.

CALL TRANSACTION 'PA20' WITHOUT AUTHORITY-CHECK AND SKIP FIRST SCREEN.
```

The inline declaration (`DATA(lv_foo) = 1.`) is correctly downported into a separate `DATA`/assignment pair — confirming the `downport` rule is running and fixing this file — but the `WITHOUT AUTHORITY-CHECK` clause on the very next statement is left untouched.

### Context
- `check_syntax` is disabled in our project config, since the data-access framework the code depends on only exists on the SAP system, not in the repo — so this isn't a `check_syntax` release-availability gap, it's specifically about `downport --fix` not touching this addition at all.
- All other clauses on the same statement (`AND SKIP FIRST SCREEN`, `USING`, ...) must be left untouched — only `WITHOUT AUTHORITY-CHECK` should be stripped.
- Workaround currently in use (CI step, outside abaplint):
  ```
  find src -name '*.abap' -print0 \
    | xargs -0 -r sed -i -E '/CALL TRANSACTION/ s/ WITHOUT AUTHORITY-CHECK//'
  ```

### Ask
Have `downport --fix` remove `WITHOUT AUTHORITY-CHECK` from `CALL TRANSACTION` statements when downporting to a syntax version where it isn't supported, the same way it already rewrites other 7.5x-only constructs (inline declarations, `NEW`, string templates, `VALUE`/table expressions, ...).

### Environment
- abaplint 2.119.48
- `syntax.version: "v702"`, `rules.downport: true`

---

fabian-werner-peras:
Closes #4297.

`WITHOUT AUTHORITY-CHECK` on `CALL TRANSACTION` is valid ABAP on every classic release (it's parsed via `verNotLang(LanguageVersion.Cloud, ...)`, not release-gated), so the statement never becomes `Unknown` under a v702 target and never reaches the existing Unknown-statement-driven downport fixes. Some v702 target systems still reject the clause at runtime/activation for reasons unrelated to the ABAP language version (see #4297 for the reporter's context and a minimal repro), so this adds a small, unconditional downport pass — following the same pattern already used for the `xsdbool(...)` and SQL `INTO`-after-`WHERE` fixes — that strips `WITHOUT AUTHORITY-CHECK` from `CALL TRANSACTION` statements, leaving every other clause (`AND SKIP FIRST SCREEN`, `USING`, `MODE`, ...) untouched. `WITH AUTHORITY-CHECK` is left alone.

## Changes
- `packages/core/src/rules/downport.ts`: new unconditional pass in `Downport.run()` plus `stripCallTransactionAuthorityCheck()`.
- `packages/core/test/rules/downport.ts`: 3 new tests (strip with trailing clauses, strip as sole clause, `WITH AUTHORITY-CHECK` left untouched).

## Testing
- `npm run compile` (tsc) — clean.
- Full suite (`mocha`, 11,005 tests) — all passing, no regressions.


larshp (approved):
thanks

larshp:
published as v2.120.54

PR #4300, opened 2026-09-16
new rule: prefer_insert_into_table

jan-jezek:
## Summary

- Adds new rule `prefer_insert_into_table` that flags `APPEND` statements and suggests replacing them with `INSERT INTO TABLE`
- Provides a quickfix for all flagged cases using token-level replacement
- `APPEND ... SORTED BY` is silently skipped (no direct single-statement equivalent)

## What's New

**Rule:** `prefer_insert_into_table`

`APPEND` always adds to the end of an internal table regardless of table type. `INSERT INTO TABLE` respects the table type — sorted and hashed tables insert at the correct position.

Covered forms (all flagged with quickfix):
- `APPEND row TO itab.` → `INSERT row INTO TABLE itab.`
- `APPEND LINES OF itab2 TO itab.` → `INSERT LINES OF itab2 INTO TABLE itab.`
- `APPEND INITIAL LINE TO itab [ASSIGNING <fs>] [REFERENCE INTO ref].`
- `APPEND row TO itab ASSIGNING <fs> [CASTING].`
- `APPEND LINES OF itab2 FROM 1 TO 3 TO itab.` (range form)

## Technical Details

The quickfix uses token-level replacement (`EditHelper.replaceToken`): replaces the `APPEND` token with `INSERT` and the last `TO` token (before the table name) with `INTO TABLE`. The "last TO" strategy correctly handles the range form where `TO` appears twice.

Rule is tagged `SingleFile`, `Styleguide`, `Quickfix` and auto-discovered via the rules index export.


larshp:
thanks, few eslint warnings

larshp (approved):

larshp:
published as v2.120.54

PR #4303, opened 2026-09-17
fix and add rule documentation links

jan-jezek:
## Summary

- `align_parameters`: update `blob/master` to `blob/main`
- `constant_classes`: fix renamed anchor `#prefer-enumeration-classes-to-constants-interfaces` → `#prefer-enum-to-constants-interfaces`
- `keyword_case`: fix renamed anchor `#use-your-pretty-printer-team-settings` → `#use-your-teams-abap-formatter-settings`
- `prefer_insert_into_table`: add specific anchor `#prefer-insert-into-table-to-append-to`; improve description to note that APPEND will dump if the sort order of a sorted table would change
- `sequential_blank`: add `extendedInformation` link to CleanABAP.md


larshp (approved):
thanks

PR #4311, opened 2026-09-19
Parse VALUE() and REFERENCE() with the identifier escape in front

### LINKED ISSUE #4308: `!VALUE(x)` in a method signature is not parsed, and the method is lost
abaplint parses `!x` and parses `VALUE(x)`, but not the two together. A
parameter written `!VALUE(x)` makes the `METHODS` / `CLASS-METHODS` statement
come back as `Unknown`, so that method is missing from
`getClassDefinition().methods`. Other methods of the class are unaffected.

In practice the form is generated rather than typed, so it covers every
parameter of every method in a class, and what arrives is a class with almost
no signatures. In two classes read off a system, abaplint parses 2 of 15
methods and 1 of 9; removing the `!` before parsing gives all of them.

`findIssues()` does report a `parser_error`, but it points at the `CLASS`
token rather than at the parameter, and a caller using the
`getClassDefinition()` API sees only a missing method.

### Reproducer

```js
const {MemoryFile, Registry} = require("@abaplint/core");
const source = `CLASS c DEFINITION PUBLIC.
  PUBLIC SECTION.
    CLASS-METHODS m1 IMPORTING !VALUE(iv_x) TYPE i.
ENDCLASS.
CLASS c IMPLEMENTATION.
ENDCLASS.`;
const reg = new Registry().addFile(new MemoryFile("c.clas.abap", source)).parse();
const obj = reg.getFirstObject();
obj.getABAPFiles()[0].getStatements().map(s => s.get().constructor.name);
// [ 'ClassDefinition', 'Public', 'Unknown', 'EndClass', 'ClassImplementation', 'EndClass' ]
obj.getClassDefinition().methods;   // []
```

### The forms, measured separately

| declaration | result |
| --- | --- |
| `IMPORTING iv_x TYPE i` | parsed |
| `IMPORTING !iv_x TYPE i` | parsed |
| `IMPORTING VALUE(iv_x) TYPE i` | parsed |
| `IMPORTING !VALUE(iv_x) TYPE i` | **`Unknown`**, method missing |
| `IMPORTING !REFERENCE(iv_x) TYPE i` | **`Unknown`**, method missing |

Lower case `!value(...)` behaves the same as the upper-case form.

Version: `@abaplint/core` 2.120.55, current on npm at the time of writing.

### Where I hit it

I read AMDP method bodies out of classes exported from a system and need the
signature to know which names in the body are IN table parameters. With this
form the signature is missing, so a body referring to `:it_ddls` — a
parameter declared three lines above it — looks like a body reading an
undeclared table variable. The message points somewhere other than the
defect, which is what made it expensive to find.

I work around it by removing the `!` before `VALUE(` prior to parsing. The
`!` escapes the identifier so it is not read as a keyword; it carries no
meaning for the interface, so removing it before parsing changes nothing
about what is read.

### Where it seems to come from

Read off the built package rather than the source, so treat it as a pointer
and not a diagnosis. The lexer emits `!VALUE` as a single `Identifier` token,
while `MethodParam` matches `VALUE` / `REFERENCE` as literal strings:

```js
// 2_statements/expressions/method_param.js
const ref   = seq("REFERENCE", tok(ParenLeft), Expressions.MethodParamName, tok(ParenRightW));
const value = seq("VALUE",     tok(ParenLeft), Expressions.MethodParamName, tok(ParenRightW));
```

and only `MethodParamName` tolerates a leading `!` — its regex is
`^!?\w*(\/\w+\/)?\w+$`. That is why `!iv_x` works and `!VALUE(iv_x)` does
not. Whether the escape belongs in those two alternatives or somewhere
earlier is yours to say; happy to send a PR if you point at the shape you
want.

Related: #2529, the same escape in front of a builtin function.

oisee:
Fixes #4308.

The lexer emits `!VALUE` as a single `Identifier` token, so a literal
`"VALUE"` in the grammar cannot match it and the whole statement falls back
to `Unknown`. `!iv_x` parses and `VALUE(iv_x)` parses; only the combination
does not.

Three expressions matched the keyword as a literal, and all three are the
same shape:

| expression | where it shows |
| --- | --- |
| `MethodParam` | `IMPORTING` / `EXPORTING` / `CHANGING` |
| `MethodDefReturning` | `RETURNING` |
| `PassByValue` | `FORM … USING` / `CHANGING` |

The third is beyond what the issue reported. I included it because it is the
same line for the same reason and leaving it would fix the place rather than
the defect — happy to drop it if you would rather keep the change to methods.

### Tests

Six cases added to the existing lists, two of them in `form.ts`:

```
CLASS-METHODS m1 IMPORTING !VALUE(iv_x) TYPE i.
CLASS-METHODS m1 IMPORTING !REFERENCE(iv_x) TYPE i.
CLASS-METHODS m1 IMPORTING !value(iv_x) TYPE i EXPORTING !value(ev_y) TYPE i.
METHODS m1 RETURNING !VALUE(rv_x) TYPE i.
METHODS m1 RETURNING !value(rv_x) TYPE i.
FORM name USING !VALUE(bar).
FORM name CHANGING !value(bar) TYPE i.
```

`RETURNING !VALUE(rv_x)` is what found the second site: it still failed after
`MethodParam` was fixed.

### Run locally

`packages/core`: `npx tsc` clean, `mocha build/test --recursive` **11060
passing, 0 failing**, `npm run lint` clean. I cannot trigger your regression
workflow from a fork, so that is the evidence I can offer.

### Where I hit it

I read AMDP method bodies out of classes exported from a system. With this
form the signature is missing, so a body referring to `:it_ddls` — a
parameter declared three lines above it — looks like a body reading an
undeclared table variable. The form is rare by file count (6 of 3052 classes
I have) but generated where it appears, so it takes a whole class with it: in
one, abaplint parses 2 of 15 methods.


larshp (changes requested):
hmm, not sure we want to tackle it like this

oisee:
Reworked. The first version fixed the parse and left the meaning wrong.

`isPassByValue()` compares the first token with `"VALUE"`, so a parameter written `!VALUE(iv_x)` parsed and then came out classified pass by **reference**: assigning to it reports `"iv_x" cannot be modified, it is readonly` on correct code. Measured on the first version:

```
VALUE(iv_x)    assignment to the parameter → no check_syntax
!VALUE(iv_x)   → check_syntax: "iv_x" cannot be modified, it is readonly
iv_x           → the same, correctly
```

So a method that used to disappear would have become a method that is quietly wrong. `isPassByValue()` is part of this change now, with a test that fails with the grammar fix alone. `!REFERENCE` needs nothing there — by reference is the default.

On `no_exclamation_escape`: it fires on this today and its quickfix removes the `!`, which is the right advice. But the rule can only tell someone to change the code — before that, abaplint is wrong *about* the code: the statement is `Unknown`, the method is missing from `getClassDefinition()`, and the `parser_error` points at `METHODS`. That is the part this PR is about; the rule keeps saying what it says afterwards.

I also looked at doing it once in `Word` instead. It is a one-line change and the suite stays green — but it is wrong: the same matcher serves the operators, so `IF 1 != 2.` starts parsing as `If` where it is `Unknown` today, and so do `!EQ` and `!<`.

Scope is narrower than the first version: method parameters and `RETURNING`. The `FORM` change is gone — `FORM f USING !p.` fails for an unrelated reason (`FormParamName` has no `!?` where `MethodParamName` does), so it wants its own change.

I checked the spelling against a system's syntax check rather than assuming it. Three runs, same class, unsaved source:

```
METHODS m1 IMPORTING !VALUE(iv_x) TYPE i.   +  iv_x = 2.   → no findings
METHODS m1 IMPORTING iv_x TYPE i.           +  iv_x = 2.   → "The field "IV_X" cannot be modified."
(a deliberate typo)                                        → "Field "..." is unknown."
```

The compiler accepts `!VALUE(...)`, `!REFERENCE(...)` and `RETURNING !VALUE(...)`, and treats the escaped one as pass by value. The third line is the control, so the empty answer above is an answer rather than a silence. Its scanner also returns `!VALUE` as a single token.

In two classes read off that system abaplint parsed 2 of 15 methods and 1 of 9; removing the `!` before parsing gives all of them.

`packages/core`: 11059 passing, 32 pending, 0 failing; eslint clean (regression doesn't run on fork branches).

There is also a middle option — a small escape-aware `str` used only by these two expressions, if you would rather it were named once rather than spelled twice. If neither shape is the one you want, say where it should go and I will move it.

larshp (approved):
fine for now

there are more larger issues in this area, but they typically dont occur in real code

